PR #41, opened 2015-10-30
Wrap input functions in callback API

lionel-:
Hello Evan,

Here is the API update discussed last week in the Haven repo. A version of the callback structure using the original input functions is included in `readstat_io_unistd.c`.

I have a limited set of test files but they all work correctly.


evanmiller:
Hi,

Thanks for the PR. A few comments:

* Can we just attach the IO handlers to the parser? Or let the IO struct be a field in the parser. That way we won't have to break existing APIs.

* I'd also like the unistd handlers to be the defaults (again to prevent breaking the API)

* Not sure why `dta_read_tag` needs both `io` and `ctx`, isn't `io` available in `ctx->io`?

* For code simplicity I think I'd prefer `io->read` etc instead of `io->read_handler` and friends

* Please make sure you are up to date with the latest ReadStat code... looking esp at https://github.com/WizardMac/ReadStat/pull/41/files#diff-e08fb30c45af0d39e7ccf214dcd9e14fL135

lionel-:
> Not sure why dta_read_tag needs both io and ctx, isn't io available in ctx→io?

It's because of initialisation timing. The DTA parser first needs to figure out if the file is xmlish, for which it uses `dta_read_tag()`, but `ctx` isn't initialised yet at this point. I tried to sort this out but couldn't find an easy way.

Thanks for your comments, I'll address them next week.

lionel-:
I'm not sure how to completely avoid an API breakage.

I can initialise a unistd `io_ctx` field by default and have it freed and replaced when the user calls `readstat_set_io_ctx()`. But we still need to specify the file name. Currently it is passed through `readstat_parse_()` functions, but this doesn't make a lot of sense with the callbacked API, since there isn't necessarily any file name (when the file is already in memory for example).

evanmiller:
I'm OK with renaming "file name" to "path" and just being a generic string that the IO handler receives. Could be a file path, a URL, or NULL.

lionel-:
Here you go.

evanmiller:
Thanks. I find the `dta_ctx_init`/`dta_ctx_preinit` distinction a little confusing. I think signatures like:

```{C}
// calls malloc
dta_ctx_t *dta_ctx_alloc(readstat_io_t *io);

// takes allocated ctx, returns 0 on success, doesn't free ctx on error
int dta_ctx_init(dta_ctx_t *ctx, int16_t nvar, int32_t nobs, unsigned char byteorder, unsigned char ds_format); 
```

Would work better. Otherwise this looks OK to me.

Are you planning to use IO handlers with the writers as well?

lionel-:
> I think signatures like:

ah yes, much better. Done in the last commit.

> Are you planning to use IO handlers with the writers as well?

Not currently. I'm not sure if something could be gained from this.

evanmiller:
> ah yes, much better. Done in the last commit.

I was actually more interested in returning `int` from `dta_ctx_init` (not just renaming the function)... otherwise it's confusing about whose responsibility it is to free. So `dta_ctx_init` should not call `dta_ctx_alloc` IMO.

> Not currently. I'm not sure if something could be gained from this.

That's fine, was just curious.

lionel-:
sorry, should be good now.

evanmiller:
Thanks! I might tweak a couple things but overall it looks good so I'll merge it in.

lionel-:
Thanks. Let me know when it's ready, I'll update the PR over at Haven.

evanmiller:
Ok I think it should be good to go now.

lionel-:
thanks!

and let me also thank you for making this library public. This is hugely appreciated.

PR #48, opened 2016-05-12
Initial GNU autotools support

jar1karp:
This a rather bare bones implementation to get started with. Autotools can be bootstrapped by running the ./autogen.sh script first. That done with some luck the usual ./configure && make && make install should work. I've only tested this on Linux (Fedora 23) though.


evanmiller:
Some notes on OS X:

* autogen.sh needs a shebang line (eg `#!/bin/bash`)
* Compiles OK but I'm not familiar with libtool -- I'd prefer a system that produces platform-specific shared libraries (dylib on  Mac, DLL on windows, etc)

jar1karp:
Ok, I've added the sheband. I though libtool is supposed to produce a proper shared library on each platform. Is it not compiling a .dylib on Mac as of now?

evanmiller:
Ah OK, it looks like that happens with "make install" and "libtool --finish". I'm a little concerned about the mess of files. Maybe this is a stupid question but do you know if there are any solutions that are simpler than autotools?

jar1karp:
Autotools are indeed rather spammy, although you only need to version those three files and do git clean -dxf . after building to tidy up. Perhaps cmake might be simpler, but it will likely get equally hairy to get the ragel steps integrated.

evanmiller:
Ok thanks. I will get this merged when I get a chance to integrate it with my own build system -- bother me in a week if I haven't moved on it.

evanmiller:
Ok, starting to play with your branch... looks like `./autogen.sh` pulls in a GPLv3 `COPYING` file, which is a problem. (ReadStat is MIT licensed.) It looks like autoreconf also requires NEWS, AUTHORS, README, and ChangeLog -- having empty files seems pointless, is there a way to work around this? I realize having versions of these files is a good idea, but we have different names for some (LICENSE vs COPYING, README.md vs README).

jar1karp:
Turns out AM_INIT_AUTOMAKE can be passed the foreign option for packages which don't follow GNU standards, which is the case here. I've pushed a commit to do this. Probably the one final thing to take care of is not installing the .la file (RPM specs typically remove them manually before collecting final binaries).

evanmiller:
Great, thanks for researching. Will keep playing with the branch and will wait for your commit re: .la files.

jar1karp:
Ok, 51a95c2 sort of takes care of the .la file. I suppose on Mac OS the runtime linker would search the default install location /usr/local/lib automatically? Typically on Linux you otherwise add ld.so config file with the non-standard path and run ldconfig (which is what the libtool --finish reminder is about).

https://www.gnu.org/software/automake/faq/autotools-faq.html#What-are-these-_002ela-files-for-and-can-I-safely-remove-them_003f

The other commits set static library disabled by default and mark ragel env variables as precious. That is, you can set/override the by running  ./configure RAGELFLAGS=-G2

evanmiller:
Ok, getting closer... is there a way to have the compilation not crap `.lo` files into the src/ directory? Would like to keep that directory clean at least.

jar1karp:
One way would be to run the build from another directory. autoreconf -i has to be run in the source directory (which contains configure.ac) but for everything else you can run configure and make from any temporary directory (basically same as with cmake).

Anyway, any temporaries can be gotten rid of easily by git clean -dxf .

evanmiller:
All right, let's move this train forward then. I'll go ahead and merge. Just as a heads up I am working on a command-line front-end to ReadStat -- I'll probably figure out how to integrate it with autotools but I might ping you with a question or two if that's OK.

Thanks for all your work on this @jar1karp!

jar1karp:
Ok, cool.

PR #51, opened 2016-05-29
Detect host OS and set additional dependencies accordingly

jar1karp:
It seems that the additional dependencies for libreadstat.so vary across platforms, i.e on Mac you need -liconv while on Linux -lm is required (for pow()).

This commit adds the detection of canonical host and then sets EXTRA_LIBS accordingly.

Also added printing the configuration results at end.


jar1karp:
Added another commit to fix the dependency specification. Building with make -j4 it would sometimes try to link the program before library was finished. This should solve such issues.

evanmiller:
Thanks!

PR #57, opened 2016-06-07
Windows, msys2 and appveyor

davidanthoff:
This is work in progress, so don't merge. This is my attempt to get things compiling and working on Windows. So far things don't work and I'm hoping I can get some help via this PR.

The build output I currently get is [this](https://ci.appveyor.com/project/davidanthoff/readstat/build/0.1.26).

As far as I can tell there are two problems:
1) something with the ``-lz`` library. Probably that just requires another dependency to be installed via pacman
2) It seems to me that things start to go wrong with various calls to ``dprintf``. I'm not sure whether those calls could be replaced with something else, or whether I'm just missing another dependency.

My real goal is to get WizardMac/DataRead.jl#12 going. For a really stable solution we would probably have to get cross-compiling on the SUSE build infrastructure going, but for starters it seemed helpful if one can compile things locally with msys2.

Please also see my comments on specific changes that I made, they need careful review, I am not really familiar with these tools :)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -33,7 +33,7 @@ endif
 
 libreadstat_la_CFLAGS = -Wall
 libreadstat_la_LIBADD = -llzma -lz @EXTRA_LIBS@
-libreadstat_la_LDFLAGS =
+libreadstat_la_LDFLAGS = -no-undefined
 
 include_HEADERS = src/readstat.h
 
--- a/configure.ac
+++ b/configure.ac
@@ -1,12 +1,13 @@
 AC_INIT([readstat], [20160511])
 AM_INIT_AUTOMAKE([foreign subdir-objects])
 AM_SILENT_RULES([yes])
-LT_INIT([disable-static])
+LT_INIT([disable-static, win32-dll])
 AC_PROG_CC
 AC_CANONICAL_HOST
 AS_CASE([$host],
 	[*darwin*], [EXTRA_LIBS="-liconv"],
 	[*linux*], [EXTRA_LIBS="-lm"],
+	[*mingw64],  [EXTRA_LIBS="-liconv"],
 	[EXTRA_LIBS=""]
 )
 AC_SUBST([EXTRA_LIBS])
--- a/.appveyor/build_script.bat
+++ b/.appveyor/build_script.bat
@@ -1,0 +1,12 @@
+set PATH=C:\msys64\usr\bin;%PATH%
+set PATH=C:\msys64\mingw64\bin;%PATH%
+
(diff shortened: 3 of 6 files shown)


Makefile.am
```diff
@@ -33,7 +33,7 @@ endif
 
 libreadstat_la_CFLAGS = -Wall
 libreadstat_la_LIBADD = -llzma -lz @EXTRA_LIBS@
-libreadstat_la_LDFLAGS =
+libreadstat_la_LDFLAGS = -no-undefined
```
davidanthoff: I'm not sure what that really does and whether this also makes sense on non-Windows platforms. Without it nothing compiled...

configure.ac
```diff
@@ -1,12 +1,13 @@
 AC_INIT([readstat], [20160511])
 AM_INIT_AUTOMAKE([foreign subdir-objects])
 AM_SILENT_RULES([yes])
-LT_INIT([disable-static])
+LT_INIT([disable-static, win32-dll])
```
davidanthoff: Same here, I'm not sure what that really does and whether this also makes sense on non-Windows platforms. Without it nothing compiled...

evanmiller:
Thanks for the PR. The autotools stuff is pretty new (to ReadStat) so I'm not too concerned about breakage at this point -- happy to merge something incomplete if you want to get it in more people's hands to test.

Regarding (2) it looks like dprintf isn't portable. I'll try replacing it with something.

davidanthoff:
I'll rebase once to trigger a travis build so that we know whether my changes break the build on unix, and then we can merge.

davidanthoff:
Alright, travis build worked, so ready to merge.

evanmiller:
Ok, Appveyor is enabled. Merging.

Thanks for seeing this through!

davidanthoff:
What is the appveyor url for this? Apparently not https://ci.appveyor.com/project/WizardMac/ReadStat.

evanmiller:
https://ci.appveyor.com/project/evanmiller/readstat

Would prefer to use the WizardMac url but I am not sure how to configure that.

davidanthoff:
http://help.appveyor.com/discussions/questions/1154-appveyor-account-for-github-organizations

davidanthoff:
And you'll have to add the badge for appveyor, one can only get at the URL for that as the appveyor account owner.

evanmiller:
Thanks. A separate account looks like too much trouble so I'll just keep the current URL for now.

Added the AppVeyor badge (currently failing on account of some iconv errors).

PR #60, opened 2016-06-07
Configure appveyor to use mingw64 instead of msys2

### LINKED ISSUE #59: Fix Windows appveyor build
Moving the discussion out of the closed PR #57.

The latest master actually build successfully on my local system with msys2, so this is probably just ta case of some missing package that we need to add via pacman.

When I build locally I also don't get the warning about ``-lz`` that shows up in the appveyor log, so trying to figure out what is missing for that might be the first step.

davidanthoff:
This fixes #59.


evanmiller:
Thanks!

PR #65, opened 2016-07-04
Fixes for cygwin

jar1karp:
Here's a couple of fixes for building on cygwin. I get linking errors due to lseek defined as lseek64 and missing libiconv dependency. This commit should take care of both. In addition, I've changed the -no-undefined flag to linker to be passed on mingw or cygwin only -- it doesn't do anything on other platforms.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -37,7 +37,7 @@ endif
 
 libreadstat_la_CFLAGS = -Wall
 libreadstat_la_LIBADD = -llzma -lz @EXTRA_LIBS@
-libreadstat_la_LDFLAGS = -no-undefined
+libreadstat_la_LDFLAGS = @EXTRA_LDFLAGS@
 
 include_HEADERS = src/readstat.h
 
--- a/configure.ac
+++ b/configure.ac
@@ -1,22 +1,28 @@
 AC_INIT([readstat], [20160511])
 AM_INIT_AUTOMAKE([foreign subdir-objects])
 AM_SILENT_RULES([yes])
-LT_INIT([disable-static, win32-dll])
+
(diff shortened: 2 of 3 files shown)


evanmiller:
Thanks!

davidanthoff:
@jar1karp You could also add building with cygwin to the appveyor testing (in addition to the msys2 build), that way you could make sure new stuff won't break things.

jar1karp:
Yep that might be useful. I'm not familiar with appveyor but I'll try to take a look some day.

davidanthoff:
I can take a stab at adding cygwin builds to appveyor.

davidanthoff:
@jar1karp I've got pretty far with the appveyor testing for cygwin, but don't know what commands I should issue to install dependencies. Do you have a) the cygwin command for install dependencies handy and b) the list of dependencies one needs to install so that readstat compiles? I think with that I should be able to finish up the appveyor testing of cygwin.

The log of the current build is [here](https://ci.appveyor.com/project/davidanthoff/readstat/build/0.1.40/job/4uraa1bmnyiuux0t).

jar1karp:
If they provide the setup tool, then installing additional packages probably works along the lines [explained here](http://help.appveyor.com/discussions/suggestions/427-pre-install-cygwin).

For cygwin64 probably something like `c:\cygwin64\setup-x86_64.exe -qnNdO -s http://cygwin.mirror.constant.com -l C:/cygwin64/var/cache/setup -P libiconv-devel -P zlib-devel -P liblzma-devel -P xz -P ragel`

davidanthoff:
Thanks, see #67.

Issue #62, opened 2016-06-14
Build Windows binaries on opensuse for julia


davidanthoff:
I'm trying to set up a build of windows binaries on build.opensuse.org. Once that is up and running we can integrated that via the WinRPM package into the julia package and have a super simple julia story for this on Windows.

I'm following the advice [here](https://github.com/JuliaLang/WinRPM.jl/issues/28#issuecomment-121844977).

My opensuse package is [here](https://build.opensuse.org/package/show/home:davidanthoff/mingw64-readstat).

But it doesn't work, if I understand the build log properly something is going wrong at the unzip stage already (maybe it doesn't even find the unzip command?). I'm sure this is just some very simple small stupid error, but I can't find it... @tkelman, any chance you could help me out here and take a brief look at my .spec file?

tkelman:
add unzip as a BuildRequires, or use the .tar.gz source file

davidanthoff:
Thanks, @tkelman, that solved that problem!

Now I'm stuck [here](https://build.opensuse.org/package/live_build_log/home:davidanthoff/mingw64-readstat/openSUSE_13.1/x86_64) that the compiler can't find iconv.h. I think that means libiconv needs to somehow be installed. I tried to just add ``libiconv`` as another BuildRequires, but that doesn't work, and it is not clear to me what to try next...

tkelman:
[mingw64-win_iconv](https://build.opensuse.org/package/show/windows:mingw:win64/mingw64-win_iconv)

davidanthoff:
And again, thanks, @tkelman, that did the trick!

The build now succeeds, but when I click "Download package" [here](https://build.opensuse.org/package/show/home:davidanthoff/mingw64-readstat) there is nothing there...

The Rpmlint results seem to complain about a noarch tag, but I don't understand what that means.

tkelman:
might be a consequence of the other warning about spec file capitalization not matching?

davidanthoff:
Alright, I'm at the point where the download rpm file has the ``readstat.exe`` file in it, so something is building an packaging.

Really I need the libreadstat-0.dll, though... I don't seem to be able to grab it, all I get is a "File not found" error.

I think there might be two avenues forward: a) why is libreadstat-0.dll not in the same folder where readstat.exe ends up? This might be an issue with the current makefiles? b) or maybe we can just figure out where it ends up, and then grab it from there...

[...]

PR #66, opened 2016-07-08
Pass -avoid-version in LDFLAGS for Cygwin and mingw

jar1karp:
This should help with #62. By passing -avoid-version mingw64 build produces libreadstat.dll (instead of libreadstat-0.dll) so DataRead.jl can actually load the dll.


evanmiller:
Thanks

PR #68, opened 2016-07-11
WIP: Link iconv statically on opensuse builds

davidanthoff:
Lets wait with merging until @tkelman and @jar1karp take a brief look at this.


tkelman:
`x86_64-w64-mingw32` is not only an opensuse name, it's what cross compilers use from other linux distros or cygwin or non-msys2 builds of mingw-w64.

tkelman:
I also suspect you'd want an i686 build to support building 32 bit dlls

davidanthoff:
Hm... I guess one question is whether anyone is ever going to cross-compile this, except for the opensuse build... I see the problem, but the other solution, a patch file, seems also not ideal to me, then we'll have to keep that updated as there are changes to configure.ac, right?

davidanthoff:
Closing because @jar1karp has a much better solution for this.

PR #69, opened 2016-07-11
Optional static linking of iconf by passing --enable-static-iconv to …

davidanthoff:
…configure

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -6,11 +6,13 @@ LT_INIT([disable-static])
 
 AC_PROG_CC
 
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -6,11 +6,13 @@ LT_INIT([disable-static])
 
 AC_PROG_CC
 
+AC_ARG_ENABLE([static-iconv], AS_HELP_STRING([--enable-static-iconv], [Link iconv statically]), [static_iconv=yes], [static_iconv=no])
```
tkelman: what's the default? my autoconf foo is rusty

evanmiller:
Thanks

PR #70, opened 2016-07-16
Autotools improvements

jar1karp:
Here's a bunch of commits to hopefully simplify the autotools process a bit. The idea is that rather than hardcoding libraries for each host we just check the availability of optional ones and explicitly check the required ones (lzma, zlib). Also, the package version (+ other info, e.g. name and url) go into config.h now. Those defines could be used whenever printing version information so I've taken them into use for the cli tool.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -36,7 +36,17 @@ if HAVE_RAGEL
 endif
 
 libreadstat_la_CFLAGS = -Wall
-libreadstat_la_LIBADD = -llzma -lz @EXTRA_LIBS@
+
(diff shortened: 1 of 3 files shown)


evanmiller:
Hi, do you know why the build is failing on Windows?

jar1karp:
Looks like iconv is not detected properly on msys2/cygwin, so build fails in linking stage. Cross-compiling on Linux using mingw64 worked fine though. Guess I'll have to rethink this.

evanmiller:
Closing, since this PR is over a year old.

Issue #29, opened 2015-07-02
SAS transport files?


markdanese:
The foreign package does them, but was wondering if they might be added at some point to this and then to haven.  
Thanks for writing such a useful library.

dpastoor:
per #23 I believe this is outside of the scope of the project (for now at least)

markdanese:
absolutely no problem.  if it ever gets elevated, here is a link to the specs, which include some C structures that might be useful.  Feel free to close this issue if that is better.
http://support.sas.com/techsup/technote/ts140.pdf
http://support.sas.com/techsup/technote/ts140_2.pdf

evanmiller:
Thanks for the links. Closing for now, but I may revisit the issue in the future.

hadley:
Seeing some enthusiasm for this file type on twitter: https://twitter.com/jetravis/status/783428520395608064, https://twitter.com/ImDataDriven/status/783425986549473282

dpastoor:
I will confirm that this is the required format for the FDA - here are links http://www.fda.gov/Drugs/DevelopmentApprovalProcess/FormsSubmissionRequirements/ElectronicSubmissions/ucm085361.htm

Currently, need to use https://github.com/MangoTheCat/sasxport

If there are any FDA related questions feel free to ask (I'm a Fellow there) 

Because of the submission process, this is also a very standard data exchange format between CRO's and pharma companies as well (along with sas7bdat, which is already handled nicely in readstat/haven)

dennisfisher:
Hadley

FDA does request files in XPT format.  If it would help, I can provide an excerpt from a boilerplate letter than they send when an NDA is submitted.  However, the SASxport package writes XPT files so further efforts may not be needed.

Dennis

Dennis Fisher MD
P < (The "P Less Than" Company)
Phone / Fax: 1-866-PLessThan (1-866-753-7784)
www.PLessThan.com <http://www.plessthan.com/>

evanmiller:
Reopening -- support for SAS Transport will be added soon.

evanmiller:
If you'd like to test out my work-in-progress code that can read and write SAS Transport, please see #87.

evanmiller:
The SAS transport implementation is more-or-less complete, therefore I'm closing this issue.

If you find bugs in the SAS transport reader or writer, please file a new bug.

PR #87, opened 2016-10-26
Support for SAS Transport format

evanmiller:
This is a work-in-progress, posting here to collect bug reports. The pull request adds read/write support for SAS Transport format, a.k.a. XPORT format, a.k.a. XPT format. New library functions:

```C
/* Read */
readstat_error_t readstat_parse_xport(readstat_parser_t *parser,
    const char *path, void *user_ctx);

/* Write */
readstat_error_t readstat_begin_writing_xport(readstat_writer_t *writer,
    void *user_ctx, long row_count);
```

Specify the preferred format version (currently, 5 or 8) with:

```C
readstat_error_t readstat_writer_set_file_format_version(readstat_writer_t *writer,
        long file_format_version);
```

The usual features (variable labels, format strings, tagged and missing values) should all work out of the box. The `readstat` binary also adds support for reading and writing the `.xpt` format, so it may be easiest to test your files this way:

    $ readstat test.xpt test.csv

Note that multi-dataset files are not supported -- the behavior when reading them is currently "undefined", i.e. will read in a bunch of garbage data after the last row of the first data set.

Please post bug reports here.


evanmiller:
Fixes #29.

PR #143, opened 2018-07-16
Fix compilation on FreeBSD

ararslan:
Currently building on FreeBSD fails since libm is required but isn't getting linked, so this adds `-lm` in the same manner as is done on Linux. With this change, compilation succeeds and all tests pass with `gmake check`.


davidanthoff:
This is great! @evanmiller, this would allow us to make things work on all the platforms we support for the julia wrapper.

evanmiller:
Thanks

PR #154, opened 2018-10-16
fixed skipping columns for XPT files

ofajardo:
When skipping columns (returning READSTAT_HANDLER_SKIP_VARIABLE from my variable handler function for certain variables) and then recovering the index after skipping with readstat_variable_get_index_after_skipping from my value handler function, the index after skipping does not look right in the case of XPT files.
It seems to me that in the function xport_process_row in readstat_xport_read.c this case is not handled.
I replaced lines 565-568 by this (copied from readstat_por_read.c line 629), and it seems to solve the issue


evanmiller:
Thanks! :-D

PR #185, opened 2019-07-12
Add API for row skipping

### LINKED ISSUE #141: Provide API to skip rows
Making it easier to read in a chunk of rows

mikmart:
Closes #141.

Implemented skipping rows at the start of the file by seeking past when possible, and doing minimal handling when not. I'm assuming this won't be very efficient for compressed data, because seeking isn't really an option (I think?), but should work well for uncompressed formats.

### DIFF
--- a/src/readstat.h
+++ b/src/readstat.h
@@ -341,6 +341,7 @@ typedef struct readstat_parser_s {
     const char             *input_encoding;
     const char             *output_encoding;
     long                    row_limit;
+    long                    rows_skip;
 } readstat_parser_t;
 
 readstat_parser_t *readstat_parser_init(void);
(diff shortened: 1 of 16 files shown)


evanmiller:
Interesting, thanks. I would prefer to name everything `row_offset` rather than `rows_skip`. I haven't tested but from a brief glance it looks good.

evanmiller:
It would be desirable to include row-skipping in the test coverage somehow.

mikmart:
Yeah, definitely. I'm not very familiar with writing tests in C but I'll look into it. When I was trying to figure this out, I was testing mostly via a haven branch in R ([this branch](https://github.com/mikmart/haven/tree/skip-testing) has the code from this PR now).

evanmiller:
The existing test suite is here:

https://github.com/WizardMac/ReadStat/tree/master/src/test

The most important files are

https://github.com/WizardMac/ReadStat/blob/master/src/test/test_readstat.c
https://github.com/WizardMac/ReadStat/blob/master/src/test/test_list.h

One approach would be to add logic so that each existing test is run a second time with an offset and a limit (e.g. offset 1 limit 1). That way we'll have coverage of all the different file formats.

mikmart:
Okay, thanks. I think I can put something together. How do I run the tests?

mikmart:
I'm having trouble compiling locally (I'm on Windows --- I get a bunch of linker errors when I try to `make`), so had to basically just use Travis to test.

Looks like the tests are passing now, just need to clean up some unused variable warnings.

evanmiller:
Excellent work, I'm glad it's all covered by tests now. I'll get this integrated and eventually make a 1.1 release out of it.

Please open a separate issue with the Windows build errors.

Thanks again!

mikmart:
Great, thanks!

mikmart:
Re: build errors. Not sure what changed, but I actually can't reproduce the problem anymore.

PR #192, opened 2019-11-16
Fix an overflow in integer literal parsing

bpfoley:
Found by Google-internal fuzzing infrastructure.

### DIFF
--- a/src/spss/readstat_sav_parse.rl
+++ b/src/spss/readstat_sav_parse.rl
@@ -1,4 +1,4 @@
-
+#include <limits.h>
 #include <stdlib.h>
 #include "../readstat.h"
 #include "../readstat_iconv.h"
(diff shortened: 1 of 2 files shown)


src/spss/readstat_sav_parse.rl
```diff
@@ -222,8 +222,13 @@ readstat_error_t sav_parse_very_long_string_record(void *data, int count, sav_ct
         }
         
         action incr_val {
-            if (fc != '\0') { 
-                temp_val = 10 * temp_val + (fc - '0'); 
+            if (fc != '\0') {
+                unsigned char digit = fc - '0';
+                if (temp_val <= (INT_MAX - digit) / 10) {
```
evanmiller: Since `temp_val` is `unsigned int` I think this should be `UINT_MAX`.

bpfoley: Good catch. Thanks.

PR #199, opened 2020-01-28
Fix ckhash

kurt-vd:
This Pull Request provides 1 fix and 1 optimization for ck_hash.
The fix is necessary in order to correctly save one of my SPSS files.


evanmiller:
Thanks!

kurt-vd:
Have you any idea when these fixes will become available in tidyverse/haven?

evanmiller:
@kurt-vd The haven merge pattern is irregular. Open an issue over there if you would like to see them.

PR #213, opened 2020-09-05
SAS dataset label

reikoch:
In addition to the name of a SAS dataset also its label is read in from sas7bdat and printed the dataset's metadata is listed.

### DIFF
--- a/src/sas/readstat_sas7bdat_read.c
+++ b/src/sas/readstat_sas7bdat_read.c
@@ -300,7 +300,10 @@ static readstat_error_t sas7bdat_parse_column_name_subheader(const char *subhead
     for (i=ctx->col_names_count-cmax; i<ctx->col_names_count; i++) {
         ctx->col_info[i].name_ref = sas7bdat_parse_text_ref(cnp, ctx);
         if (i == 0) {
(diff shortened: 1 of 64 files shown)


src/sas/readstat_sas7bdat_read.c
```diff
@@ -298,6 +300,19 @@ static readstat_error_t sas7bdat_parse_column_name_subheader(const char *subhead
...
+            } else off = ctx->u64 ? 36 : 12;
+            if (ctx->col_info[0].name_ref.offset > ctx->text_blob_lengths[0]) {
+                retval = READSTAT_ERROR_PARSE;
+                goto cleanup;
+            }
+            memcpy(ctx->file_label,
+                   &ctx->text_blobs[0][off],
+                   ctx->col_info[0].name_ref.offset - off
+                   );
+        }
```
evanmiller, line 315: I think additional pointer validation is necessary here. We need to ensure:

* The first `memcmp` does not overrun `text_blob_lengths[0]`
* `off` is not larger than `ctx->col_info[0].name_ref.offset`
* `ctx->file_label` is not overrun during the `memcpy`

Additionally: `file_label` likely needs to be recoded. So I suggest using `readstat_convert` here instead of the `memcpy`. That will automatically solve the third issue. Something like

```C
readstat_convert(ctx->file_label, sizeof(ctx->file_label),
    &ctx->text_blobs[0][off], ctx->col_info[0].name_ref.offset - off, ctx->converter)
```

Then check the return value.

reikoch: Should be addressed with the last commit

PR #216, opened 2020-09-22
Fix Windows unicode paths

zebrys:
Fix the issue #200 (on Windows).

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -128,6 +128,7 @@ noinst_HEADERS = \
        src/bin/write/mod_xlsx.h \
        src/bin/write/module.h \
        src/bin/write/module_util.h \
+       src/bin/util/main.h \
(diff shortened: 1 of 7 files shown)


evanmiller:
Do we need `portable_main` applied to the `extract_metadata` CLI tool also?

zebrys:
> Do we need `portable_main` applied to the `extract_metadata` CLI tool also?

@evanmiller Are you talking about the replacement of `main` by `portable_main` in src/bin/extract_metadata.c? This is already the case.
Overall, all programs that take possible unicode parameters in input (eg. filenames) should use `portable_main` (only readstat and extract_metadata seems to need that I guess).

evanmiller:
Ah yes I was reading the patch too quickly. I will merge it in.

Issue #220, opened 2020-12-14
Time/Duration format in extract_metadata


basgys:
Hi WizardMac team!

First of all, thank you very much for your hard work on this open source project. 

I've started to build a tool to read SPSS files and I have a problem with time/duration columns. I use readstat/extract_metadata and `extract_metadata` extracts time column as simple integer. However, when I use WizardMac, it recognises the time column. (See screenshot)

## Our tool vs WizardMac

<img width="1072" alt="Screenshot 2020-11-24 at 15 07 06" src="https://user-images.githubusercontent.com/491123/102067254-22829180-3dfb-11eb-9f9a-d1a7ab4ae3da.png">

## extract_metadata output

I tested with both 1.1.4 and 1.1.5.

```json
{
  "type": "SPSS",
  "variables": [
    {
      "type": "DATE",
      "name": "Date1",
      "label": "Date format 1"
    },
    {
      "type": "DATE",
      "name": "Date2",
      "label": "Date format 2"
    },
    {
      "type": "NUMERIC",
      "name": "Heure",
      "label": "Date - heure/seconde"
    },
    {
      "type": "STRING",
      "name": "Texte",
      "label": "Blabla"
    }
  ]
}
```

## Source file

[date.sav.zip](https://github.com/WizardMac/ReadStat/files/5687996/date.sav.zip)

## Time-related PRs

1. https://github.com/WizardMac/ReadStat/issues/211
1. https://github.com/WizardMac/ReadStat/issues/155 (Has this been implemented already?)
1. https://github.com/WizardMac/ReadStat/issues/91

## Metadata

Besides time/duration, is there another known data type currently not extracted?

Cheers!

evanmiller:
Hi, thanks for the report. The ReadStat library is time-agnostic and simply returns raw numbers alongside the format string – it's up to the client program to make sense of "DATE5", etc. Different formats use different conventions - SAS and Stata use seconds since 1960, certain Stata formats use days since 1960, and SPSS uses seconds since the advent of the Gregorian calendar in 1582 (!). ReadStat is essentially silent on these issues.

I believe `extract_metadata` has some additional time facilities – but this tool was contributed by someone else, and I don't often use it so I can't shed much light without digging into the source code. As you discovered, Wizard has extensive logic to process time, date, and duration formats correctly, but this logic is not in the main library or in the CLI tools.

Part of that logic exists in a separate library here:

https://github.com/WizardMac/TimeFormatStrings

[...]

Issue #221, opened 2020-12-14
extract_metadata error: not a number: d


basgys:
I've attached a valid STATA file to this PR that crashes on both `extract_metadata` and WizardMac. Unfortunately, I cannot really give you more information, as the error is rather opaque. 

## extract_metadata output

Tested with both 1.1.4 and 1.1.5.

```
parsing dta file
pass 1 done
parsing dta file
extracting value labels for IDNUM
extracting value labels for PARTICIP
src/bin/extract_metadata.c:59 not a number: d
```

## WizardMac error

<img width="472" alt="Screenshot 2020-12-14 at 11 18 06" src="https://user-images.githubusercontent.com/491123/102069600-47c4cf00-3dfe-11eb-91c5-a55f1054f42a.png">

## Test files

[files.zip](https://github.com/WizardMac/ReadStat/files/5688112/files.zip)

Note that this file has been converted to SPSS and it works.

evanmiller:
Thanks, I'll look at this later today.

evanmiller:
These look like two separate errors. I realize the error messages aren't especially helpful.

The first error is a shortcoming in `extract_metadata`, which only recognizes one Stata date format (`%td`). As you may have gathered by now, it's not a particularly robust tool. I can probably add a couple of workarounds to get you going in the right direction.

The second error (in Wizard) is an encoding error of some kind processing the DTA file. The problem also shows up, without an error message, in the SPSS file. Here's a screenshot of the SPSS file, column `a1_foccu` if you're following at home:

![image](https://user-images.githubusercontent.com/134711/102093698-77200f80-3def-11eb-8cad-1677dfeda6c1.png)

My guess here is that this file uses a system encoding that is not reported in the DTA file, probably Latin-1. Whatever program converted it to SPSS happily crunched through the bad encoding, but ReadStat will refuse. By the way, the file encoding can be supplied manually to the ReadStat library, but the CLI tools do not support this option.

basgys:
Hi Evan!

Thanks for spending time looking into this issue.

Yes, `extract_metadata` is indeed not that robust. However, both `readstat` and `extract_metadata` are saving us a lot of time in our project. Our other option is to use `cgo` to import the readstat package as a library in our Go project, but it would be a significant increase in development time.

Is there a chance to include more flags to the `readstat` CLI to support encoding? Because if `readstat` cannot open certain files, I am not exactly sure what use cases it covers.

[...]

PR #223, opened 2020-12-16
implements duration support in extract_metadata

basgys:
Hi @evanmiller,

I made a first implementation to support duration in `extract_metadata`. This feature would break backward compatibility, so it is up to you to tell me if we should use a different strategy.

I use this website as a reference for the existing data types https://libguides.library.kent.edu/SPSS/DatesTime.

## Other option

As I mentioned on the PR #220, the other option would be to label the type as `NUMERIC` for both dates and durations, but then include a format field in the JSON payload that gives the format:

    TIME8 => hh:mm:ss

```json
{
   "type": "NUMERIC",
   "format": "TIME",
   "pattern": "hh:mm:ss"
}
```

Possible formats:

1. NUMBER
1. PERCENT
1. CURRENCY
1. DATE
1. TIME
1. DATE_TIME
1. UNSPECIFIED

## Note on data type handling

I added the following comment to the code:

        // All-or-nothing is probably not the best strategy for data type extraction.
        // When SPSS/STATA introduces new types, metadata extraction could fail.
        // It would be wiser to simply label the field as "UNKNOWN".

It is a philosophy I guess, but I would argue that this error should be handled further up in the stack. `extract_metadata` should be responsible to, as its name states, extracting metadata. Now if a new type is introduced, it should simply return the column as categorised instead of failing. Perhaps this new column is not important at all and does not deserve a complete failure in the extraction process.

## Style

Also, note that I haven't worked with C in a long time and I did not want to start going crazy with the refactor, but I'm pretty sure we can't have a more elegant function than this chain of `strncmp`. What would you suggest?

In Go I would have used a map of strings:

```go
  var categories = map[string]string{
    "TIME8": "DURATION",
    "DATE9": "DATE",
    // ...
  }
```

### DIFF
--- a/src/bin/extract_metadata.c
+++ b/src/bin/extract_metadata.c
@@ -14,12 +14,6 @@
 #include "write/json/write_missing_values.h"
 #include "write/json/write_value_labels.h"
 
-typedef enum extract_metadata_type_e {
-    EXTRACT_METADATA_TYPE_NUMERIC,
-    EXTRACT_METADATA_TYPE_STRING,
-    EXTRACT_METADATA_TYPE_UNKNOWN
-} extract_metadata_type_t;
-
 static const char* extract_metadata_type_str(extract_metadata_type_t t) {
     switch (t) {
      case EXTRACT_METADATA_TYPE_NUMERIC:
@@ -31,16 +25,6 @@ static const char* extract_metadata_type_str(extract_metadata_type_t t) {
     }
     return "UNKNOWN";
 }
-
-typedef enum extract_metadata_format_e {
-    EXTRACT_METADATA_FORMAT_NUMBER,
-    EXTRACT_METADATA_FORMAT_PERCENT,
-    EXTRACT_METADATA_FORMAT_CURRENCY,
-    EXTRACT_METADATA_FORMAT_DATE,
-    EXTRACT_METADATA_FORMAT_TIME,
-    EXTRACT_METADATA_FORMAT_DATE_TIME,
-    EXTRACT_METADATA_FORMAT_UNSPECIFIED
-} extract_metadata_format_t;
 
 static const char* extract_metadata_format_str(extract_metadata_format_t format) {
     switch (format) {
--- a/src/bin/extract_metadata.c
+++ b/src/bin/extract_metadata.c
@@ -121,115 +121,145 @@ static int handle_variable_sav(int index, readstat_variable_t *variable, const c
 
         // Extract format
         // SPSS data types: https://libguides.library.kent.edu/SPSS/DatesTime
-        // Pattern formats: https://developers.google.com/sheets/api/guides/formats
+        // Pattern formats: https://unicode.org/reports/tr35/tr35-dates.html#Date_Format_Patterns
         // TODO: Extract currency
         if (vformat) {
             if (hasPrefix(vformat, "DATE9") == 0) {
+                // e.g. 31-JAN-13
                 format = EXTRACT_METADATA_FORMAT_DATE;
-                pattern = "d-m-y";
+                pattern = "dd-MMM-yy";
             } else if (hasPrefix(vformat, "DATE11") == 0) {
+                // e.g. 31-JAN-13
                 format = EXTRACT_METADATA_FORMAT_DATE;
-                pattern = "dd-m-yyyy+";
+                pattern = "dd-MMM-yyyy";
             } else if (hasPrefix(vformat, "ADATE8") == 0) {
+                // e.g. 01/31/13
                 format = EXTRACT_METADATA_FORMAT_DATE;
-                pattern = "m/d/y";
+                pattern = "MM/dd/yy";
             } else if (hasPrefix(vformat, "ADATE10") == 0) {
+                // e.g. 01/31/2013
                 format = EXTRACT_METADATA_FORMAT_DATE;
-                pattern = "m/d/yyyy+";
+                pattern = "MM/dd/yyyy";
             } else if (hasPrefix(vformat, "EDATE8") == 0) {
+                // e.g. 31.01.13
                 format = EXTRACT_METADATA_FORMAT_DATE;
-                pattern = "d.m.y";
+                pattern = "dd.MM.yy";
             } else if (hasPrefix(vformat, "EDATE10") == 0) {
+                // e.g. 31.01.2013
                 format = EXTRACT_METADATA_FORMAT_DATE;
-                pattern = "d.m.yyyy+";
+                pattern = "dd.MM.yyyy";
             } else if (hasPrefix(vformat, "SDATE8") == 0) {
+                // e.g. 13/01/31
                 format = EXTRACT_METADATA_FORMAT_DATE;
-                pattern = "y/m/d";
+                pattern = "yy/MM/dd";
             } else if (hasPrefix(vformat, "SDATE10") == 0) {
+                // e.g. 2013/01/31
                 format = EXTRACT_METADATA_FORMAT_DATE;
-                pattern = "yyyy+/m/d";
+                pattern = "yyyy/MM/dd";
             } else if (hasPrefix(vformat, "DATETIME17") == 0) {
+                // e.g. 31-JAN-2013 01:02
                 format = EXTRACT_METADATA_FORMAT_DATE_TIME;
-                pattern = "d-mm-yyyy+ h:mm";
+                pattern = "dd-MMM-yyyy hh:mm";
             } else if (hasPrefix(vformat, "DATETIME20") == 0) {
+                // e.g. 31-JAN-2013 01:02:33
                 format = EXTRACT_METADATA_FORMAT_DATE_TIME;
-                pattern = "d-mm-yyyy+ h:mm:ss";
+                pattern = "dd-MMM-yyyy hh:mm:ss";
             } else if (hasPrefix(vformat, "DATETIME23.2") == 0) {
+                // e.g. 31-JAN-2013 01:02:33.72
                 format = EXTRACT_METADATA_FORMAT_DATE_TIME;
-                pattern = "d-mm-yyy+ h:mm:ss";
+                pattern = "dd-MMM-yyyy hh:mm:ss.SS+";
             } else if (hasPrefix(vformat, "YMDHMS16") == 0) {
+                // e.g. 2013-01-31 1:02
                 format = EXTRACT_METADATA_FORMAT_DATE_TIME;
-                pattern = "yyyy+-m-d h:mm";
+                pattern = "yyyy-MM-dd h:mm";
             } else if (hasPrefix(vformat, "YMDHMS19") == 0) {
+                // e.g. 2013-01-31 1:02:33
                 format = EXTRACT_METADATA_FORMAT_DATE_TIME;
-                pattern = "yyyy+-m-d h:mm:ss";
+                pattern = "yyyy-MM-dd h:mm:ss";
             } else if (hasPrefix(vformat, "YMDHMS19.2") == 0) {
+                // e.g. 2013-01-31 1:02:33.72
                 format = EXTRACT_METADATA_FORMAT_DATE_TIME;
-                pattern = "yyyy+-m-d h:mm:ss";
+                pattern = "yyyy-MM-dd h:mm:ss.SS+";
             } else if (hasPrefix(vformat, "MTIME5") == 0) {
+                // e.g. 1754:36
                 format = EXTRACT_METADATA_FORMAT_TIME;
                 pattern = "[m+]:[s+]";
             } else if (hasPrefix(vformat, "MTIME8.2") == 0) {
+                // e.g. 1754:36.58
                 format = EXTRACT_METADATA_FORMAT_TIME;
                 pattern = "[m+]:[s+]";
             } else if (hasPrefix(vformat, "TIME5") == 0) {
+                // e.g. 29:14
                 format = EXTRACT_METADATA_FORMAT_TIME;
                 pattern = "[h+]:[m+]";
             } else if (hasPrefix(vformat, "TIME8") == 0) {
+                // e.g. 29:14:36
                 format = EXTRACT_METADATA_FORMAT_TIME;
                 pattern = "[h+]:[m+]:[s+]";
             } else if (hasPrefix(vformat, "TIME11.2") == 0) {
+                // e.g. 29:14:36.58
                 format = EXTRACT_METADATA_FORMAT_TIME;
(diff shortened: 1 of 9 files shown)


evanmiller:
Hi, thanks for the contribution. I don't have a strong opinion on the type vs format debate – I would suggest however looking at the `readstat` utility to ensure it can still use the JSON output of `extract_metadata` to read in CSV. Reverse-formatting is a hard problem, so I don't necessarily expect a full solution.

Re: C style I will add some comments in a code review.

src/bin/extract_metadata.c
```diff
@@ -69,21 +69,41 @@ static int handle_variable_sav(int index, readstat_variable_t *variable, const c
...
-                    strncmp(format, "ADATE", sizeof("ADATE")-1) == 0 || 
-                    strncmp(format, "EDATE", sizeof("EDATE")-1) == 0 || 
+        break;
+    case READSTAT_TYPE_CLASS_NUMERIC:
+        // Check for date and duration types
+        // https://libguides.library.kent.edu/SPSS/DatesTime
+        if (format && (strncmp(format, "DATE", sizeof("DATE")-1) == 0 ||
+                    strncmp(format, "ADATE", sizeof("ADATE")-1) == 0 ||
+                    strncmp(format, "EDATE", sizeof("EDATE")-1) == 0 ||
                     strncmp(format, "SDATE", sizeof("SDATE")-1) == 0)) {
```
evanmiller, line 82: Since `format` is a regular C string, `strcmp` instead of `strncmp` will be fine here.

basgys:
> I would suggest however looking at the readstat utility

Of course! Could you point me to this utility? And regarding a potential break in backward compatibility, what do you think?

Cheers!

src/bin/extract_metadata.c
```diff
@@ -69,21 +69,41 @@ static int handle_variable_sav(int index, readstat_variable_t *variable, const c
...
+            // TODO: Include all remaining date types
             type = "DATE";
+        } else if (format && (strncmp(format, "MTIME5", sizeof("MTIME5")-1) == 0 ||
+                    strncmp(format, "MTIME8.2", sizeof("MTIME8.2")-1) == 0 ||
+                    strncmp(format, "TIME5", sizeof("TIME5")-1) == 0 ||
+                    strncmp(format, "TIME8", sizeof("TIME8")-1) == 0 ||
+                    strncmp(format, "TIME11.2", sizeof("TIME11.2")-1) == 0 ||
+                    strncmp(format, "DTIME9", sizeof("DTIME9")-1) == 0 ||
+                    strncmp(format, "DTIME12", sizeof("DTIME12")-1) == 0 ||
+                    strncmp(format, "DTIME15.2", sizeof("DTIME15.2")-1) == 0)) {
```
evanmiller, line 92: You might simplify these comparisons with `sscanf`, which can do basic pattern matching. I typically do something like

```C
long width;
if (sscanf(format, "MTIME%ld", &width) == 1) {
} else if (sscanf(format, "TIME%ld", &width) == 1) {
...
```

It's probably okay to be a little loose with the allowed widths and precisions.

src/bin/extract_metadata.c
```diff
@@ -69,21 +69,41 @@ static int handle_variable_sav(int index, readstat_variable_t *variable, const c
...
             type = "NUMERIC";
             decimals = extract_decimals(format, 'F');
         }
-    } else {
+        break;
+    default:
+        // All-or-nothing is probably not the best strategy for data type extraction.
+        // When SPSS/STATA introduces new types, metadata extraction could fail.
+        // It would be wiser to simply label the field as "UNKNOWN".
+        // TODO: Removes this failure?
```
evanmiller, line 103: I'm okay being strict. It often brings people here to file bug reports :-)

evanmiller:
> > I would suggest however looking at the readstat utility
> 
> Of course! Could you point me to this utility? And regarding a potential break in backward compatibility, what do you think?
> 
> Cheers!

This function would need to be modified for instance:

https://github.com/WizardMac/ReadStat/blob/f37a744fac93e4503e2749e3a8b27d0e10bfd6f3/src/bin/read_csv/json_metadata.c#L188

The CSV reading code (part of the main `readstat` program) is here:

https://github.com/WizardMac/ReadStat/tree/dev/src/bin/read_csv

I'm fine breaking backward compatibility but for the sake of politeness this would probably need to be documented and wait until a 1.2 release.

basgys:
@evanmiller I've implemented format and pattern on `extract_metadata`.

If you are happy with this solution, I will updates readtstat CSV export to implement those new formats.

I also included STATA date/datetime as requested here: https://github.com/WizardMac/ReadStat/issues/221#issuecomment-747954570

evanmiller:
Are the Google Sheets formats standard in any way?

I would strongly suggest looking closely at the [TimeFormatStrings](https://github.com/WizardMac/TimeFormatStrings) library I mentioned in another thread - this solves the exact problem of parsing and converting SPSS / Stata/ SAS format strings, and can convert them reliably into a [UTS-35 Date Format Pattern](https://unicode.org/reports/tr35/tr35-dates.html#Date_Format_Patterns) or into an Excel format. The library currently lacks numeric and duration support, but I think already it would simplify your existing code and open up some other possibilities for `extract_metadata` to offer output format options.

Depending on your commitment to this, probably the best solution would be to move some of this logic into the core ReadStat library – i.e. add a `READSTAT_TYPE_DATE` and `READSTAT_TYPE_DATETIME` type, and return standardized date values, as well as standardized format strings. That way `extract_metadata` could be relatively thin. But it would be a significant amount of work, which is why I've been putting it off for a long time!

basgys:
> Are the Google Sheets formats standard in any way?

I took something popular enough as a starting point, but I'd be more than happy to use the [Unicode date format pattern](https://unicode.org/reports/tr35/tr35-dates.html#Date_Format_Patterns) if it covers our needs. Shall we use that?

> Depending on your commitment to this.

Perhaps before continuing this work, it would be good to clarify where we want to go. I would suggest not to go down the rabbit hole with an endless rework. We can start with small incremental improvements.

> the best solution would be to move some of this logic into the core ReadStat library

If you want to include more logic into the code ReadStat, I would be happy to help refactoring `extract_metadata` to simplify this code. I believe using a library like [json-c](the best solution would be to move some of this logic into the core ReadStat library would also help to make the code more robust. 

However, as I said above, I don't want to rework the whole library alone. I don't feel comfortable enough with the code base and `C` in general. Also, I don't have enough time on this project anyway.

If you are happy with it, I will do the following changes:

1. Use UTS-35 date format pattern
2. Refactor CSV export

evanmiller:
Yes the rabbit hole is deep in this one. I'm fine with the current solution of leaving dates as Numeric and indicating their format in a separate field as you do. I would prefer UTS-35 over Google Sheets since it's a Unicode standard – but I'm not sure how (or if) UTS-35 handles durations so you might need to check that first.

evanmiller:
>  I believe using a library like json-c would also help to make the code more robust.

I've had good experiences with [YAJL](https://github.com/lloyd/yajl/issues), but json-c appears to be better maintained.

basgys:
> I would prefer UTS-35 over Google Sheets since it's a Unicode standard – but I'm not sure how (or if) UTS-35 handles durations so you might need to check that first.

I will use UTS-35 for dates and if we can't represent duration, I will use the Google Sheets standard. It will be our dialect of UTS-35.

basgys:
> I've had good experiences with YAJL, but json-c appears to be better maintained.

I don't have any strong opinion on that matter. I just think concatenating strings to output a JSON payload is waiting for a disaster to happen (to be slightly dramatic).

evanmiller:
> I don't have any strong opinion on that matter. I just think concatenating strings to output a JSON payload is waiting for a disaster to happen (to be slightly dramatic).

It would be great to use a real JSON library on the CSV reading side as well.

basgys:
> > I don't have any strong opinion on that matter. I just think concatenating strings to output a JSON payload is waiting for a disaster to happen (to be slightly dramatic).
> 
> It would be great to use a real JSON library on the CSV reading side as well.

I am not quite sure to understand why the CSV reading has to work with JSON. I assume this would be removed when most of the metadata extraction will be transferred from `extract_metadata` to `readstat`.

evanmiller:
> I am not quite sure to understand why the CSV reading has to work with JSON. I assume this would be removed when most of the metadata extraction will be transferred from `extract_metadata` to `readstat`.

The CSV reader uses an `extract_metadata`-compatible JSON file to define the output schema, as described in the README here

https://github.com/WizardMac/ReadStat#command-line-usage-with-csv-input

src/bin/read_csv/mod_dta.c
```diff
@@ -190,14 +190,39 @@ void produce_missingness_dta(void *csv_metadata, const char* column) {
...
+            var->type = READSTAT_TYPE_DOUBLE;
+            snprintf(var->format, sizeof(var->format), "%%9.%df", get_decimals(c->json_md, column));
+        break;
+        case EXTRACT_METADATA_FORMAT_CURRENCY:
+            var->type = READSTAT_TYPE_DOUBLE;
+            snprintf(var->format, sizeof(var->format), "%%9.%df", get_decimals(c->json_md, column));
+        break;
+        case EXTRACT_METADATA_FORMAT_DATE:
+            var->type = READSTAT_TYPE_INT32;
+            snprintf(var->format, sizeof(var->format), "%s", "%td");
```
basgys, line 211: @evanmiller Is it supposed to be `%d` for dates, `%td` for date times, and `%t` for times?

src/bin/read_csv/mod_sav.c
```diff
@@ -125,14 +125,39 @@ void produce_missingness_sav(void *csv_metadata, const char* column) {
...
+            var->type = READSTAT_TYPE_DOUBLE;
+            snprintf(var->format, sizeof(var->format), "F8.%d", get_decimals(c->json_md, column));
+        break;
+        case EXTRACT_METADATA_FORMAT_CURRENCY:
+            var->type = READSTAT_TYPE_DOUBLE;
+            snprintf(var->format, sizeof(var->format), "F8.%d", get_decimals(c->json_md, column));
+        break;
+        case EXTRACT_METADATA_FORMAT_DATE:
+            var->type = READSTAT_TYPE_DOUBLE;
+            snprintf(var->format, sizeof(var->format), "%s", "EDATE40");
```
basgys, line 146: @evanmiller Same as .dta, should we include other formats for date time and time?

basgys (commented):
Changelog:

   1. Use UTS-35 date format pattern
   2. Refactor CSV export
   
As a general feedback, I don't have SPSS nor STATA, so unfortunately I can't really test all those changes. I would be great if you could run some tests on your end @evanmiller. Let me know what you think.

src/bin/read_csv/mod_csv.c
```diff
@@ -17,13 +18,39 @@ rs_read_module_t rs_read_mod_csv = {
...
+    extract_metadata_type_t coltype = column_type(c->json_md, column, c->output_format);
+    switch (coltype) {
+    case EXTRACT_METADATA_TYPE_NUMERIC:;
+        extract_metadata_format_t colformat = column_format(c->json_md, column);
+        switch (colformat) {
+        case EXTRACT_METADATA_FORMAT_NUMBER:
+            var->type = READSTAT_TYPE_DOUBLE;
+        break;
+        case EXTRACT_METADATA_FORMAT_PERCENT:
+            var->type = READSTAT_TYPE_STRING;
```
evanmiller, line 30: Is this supposed to be READSTAT_TYPE_DOUBLE?

evanmiller (commented):
In addition to in-line comments, are you going to parse date (and duration) strings or expect raw numbers (intervals since epoch) to be supplied in the CSV file? I would also like to see some documentation added to the README about how the new features will work.

basgys:
@evanmiller Hi Evan! Sorry for the delay, I just got busy on other projects.

I updated based on your last feedback. Let me know if I missed something, otherwise I think we are ready to merge.

evanmiller:
@basgys Thanks. Please point this PR at the `dev` branch (`master` is considered stable) and I will get this merged in.

basgys:
@evanmiller done

evanmiller:
@basgys Thanks. I believe this file will also need updating since the JSON schema has changed:

https://github.com/WizardMac/ReadStat/blob/master/variablemetadata_schema.json

basgys:
> @basgys Thanks. I believe this file will also need updating since the JSON schema has changed:
> 
> https://github.com/WizardMac/ReadStat/blob/master/variablemetadata_schema.json

I updated the schema. I am not very familiar with this standard, so I hope it reflects the output. Otherwise, I would probably need help on this.

evanmiller:
@basgys Thanks. I'm not too familiar with it either, but I just wanted to make sure things weren't falling out of date. I guess we should get the schema under test coverage somehow, but that will be its own project.

The CIFuzzer indicates that there's a new crash introduced by this PR. I will need to investigate it before merging.

evanmiller:
The crash turned out to be unrelatedm so I'll get this merged in. Thanks for your patience!

basgys:
@evanmiller I just tried to compile the development branch and got the following error:

```
src/bin/read_csv/read_csv.c: In function 'produce_column_header':
src/bin/read_csv/read_csv.c:35:41: error: 'METADATA_COLUMN_TYPE_DATE' undeclared (first use in this function)
   35 |     c->is_date[c->columns] = coltype == METADATA_COLUMN_TYPE_DATE;
      |                                         ^~~~~~~~~~~~~~~~~~~~~~~~~
src/bin/read_csv/read_csv.c:35:41: note: each undeclared identifier is reported only once for each function it appears in
make: *** [Makefile:3424: src/bin/read_csv/readstat-read_csv.o] Error 1
```

Did you merge the bug fix?

evanmiller:
@basgys What is "the bug fix"?

basgys:
> @basgys What is "the bug fix"?

When you said

> The crash turned out to be unrelated so I'll get this merged in. Thanks for your patience!

I thought you meant you had to fix something before merging the PR. My bad.

Does this branch compiles for you? Or do you have a similar error?

evanmiller:
I wasn't compiling with CSV support so I didn't see this error. Should be fixed in `dev` now!

basgys:
> I wasn't compiling with CSV support so I didn't see this error. Should be fixed in `dev` now!

It crashes again. `EXTRACT_METADATA_TYPE_DATE` does not exist anymore. I replaced it with 

```c
    extract_metadata_format_t colformat = column_format(c->json_md, column);
    c->is_date[c->columns] = colformat == EXTRACT_METADATA_FORMAT_DATE;
``

See PR https://github.com/WizardMac/ReadStat/pull/230

PR #224, opened 2020-12-20
Update NEWS for 1.1.5

NilsEnevoldsen:
Looks like 1.1.5 was missing its NEWS. This should be reviewed for accuracy, as it's based only on my reading of the commit messages.


evanmiller:
Thanks! The release notes here might be slightly more comprehensive:

https://github.com/WizardMac/ReadStat/releases/tag/v1.1.5

NilsEnevoldsen:
Oh, that also works. You can close this if you want and add those notes instead.

evanmiller:
> Oh, that also works. You can close this if you want and add those notes instead.

Will do, thanks for taking note!

PR #230, opened 2021-01-20
Fixes build caused by invalid coltype

basgys:


basgys:
@evanmiller I just wanted to let you know that I've tested this branch and it works now.

PR #242, opened 2021-05-12
Fix paths to make them consistent between x86 and x64 target architectures

zebrys:
Hello,

We found that the Visual Studio project was configured with inconsistent paths while building ReadStat in 64-bits (we have always used x86 until now). This PR make the paths more uniform (one directory for x86 and one for x64).


evanmiller:
Thanks!

PR #252, opened 2021-08-19
Keep invalid multi-byte sequences unedited in iconv encoding conversions

gorcha:
Hi @evanmiller,

This PR fixes the issue in tidyverse/haven#615.

To summarise: SPSS allows mixed encodings in files, so for e.g. we have seen SPSS files in the wild that are encoded in UTF-8 but have a handful of isolated characters encoded in Latin-1. ReadStat converts all strings using iconv, and when iconv hits a Latin-1 encoded character it sees it as an invalid multi-byte sequence, throws an EILSEQ error, and ReadStat throws a READSTAT_ERROR_CONVERT_BAD_STRING and aborts.

This PR changes the behaviour when EILSEQ is encountered - now the offending character is copied over unedited and the iconv conversion continues from the next character.

In previous ReadStat versions this issue probably didn't show up so much, since files that were already in the target encoding were not processed by iconv until ReadStat 1.1.4, but it seems to be relatively common.

Just a note on the final "else" statement for "unknown error codes" - this is to deal with the test file linked at the top of Roche/pyreadstat#67, where iconv throws an error but doesn't set any of the known error codes.
Finishing the conversion without throwing an error when an unknown error code is encountered matches the current ReadStat behaviour.


evanmiller:
Thanks for the contribution. I see a few different possible ways of handling invalid byte sequences. ReadStat's current behavior of rejecting the whole file has its detractors, but it does provide a strong guarantee to value handlers that the data will be clean and valid UTF-8 that accurately represents the data in the file.

Your proposal copies the "bad bytes" straight to the output, while continuing to convert the "good bytes". An issue with this approach is that a single output string may have mixed encoding. By default, the "good bytes" will be converted to UTF-8, with the "bad bytes" in the same string remaining unconverted. It will be difficult for any software to recover the correct string from this.

Some alternative behaviors in the face of invalid sequences would be:
* Copy over the entire original byte sequences
* Hand off the byte sequences to a special handler for conversion
* Skip over each bad byte
* Insert a Unicode replacement character � in place of each bad byte

It might be that the choice of behavior should be customizable. The calling program at least should have some indication that data loss may have occurred in a particular cell (perhaps a special flag set in `readstat_value_t`).

I will leave this pull request open but it is my opinion that a more comprehensive solution is needed for this problem.

evanmiller:
A few workarounds that haven could implement without changes to ReadStat:

```C
readstat_set_handler_character_encoding(parser, NULL); // Pass literal bytes to value handler
readstat_set_handler_character_encoding(parser, "UTF-8//IGNORE"); // Skip illegal byte sequences
readstat_set_handler_character_encoding(parser, "UTF-8//TRANSLIT"); // Insert replacement character
```

gorcha:
That all makes sense, thanks for the detailed explanation. It seems to me like the best approach might be to add a "bad byte" handler to the callbacks in the parser, then use that to process bad bytes in `readstat_convert()` (e.g. in place of the `memcpy()` call in the EILSEQ block)?

Happy to have a go at it if you don't mind reviewing :slightly_smiling_face:

I had a go at using "UTF-8//IGNORE" and "UTF-8//TRANSLIT" in the meantime but no luck unfortunately - iconv still throws the EILSEQ error since they're invalid bytes in the input encoding. Using NULL works but only because it skips the conversion (since there's no destination charset).

Thanks for the feedback and suggested workarounds, very much appreciated!

evanmiller:
Thanks for the update, didn't realize IGNORE and TRANSLIT would fail in the face of illegal byte sequences. (Makes sense esp for multi-byte encodings.)

I'm hesitant to add more handlers, but maybe a "bad sequence" handler would let programs keep their own logs of failed conversions and give end users some recovery options.

gorcha:
I've added in the bad byte handler here, with a few example handler functions in readstat_convert.c.
If no handler is provided it'll follow the current behaviour (return a `READSTAT_ERROR_CONVERT_BAD_STRING`).

A couple of things to note:
* This allows custom handling of the bad bytes, but I'm not sure how useful this will be for keeping logs of failed conversions, since `readstat_convert()` doesn't have access to any contextual information (e.g. where the string came from). It seems to me like this would need a bit of a rewrite.
* It feels a bit clunky passing the bad byte handler into `readstat_convert()` every time. Had a thought that it might be cleaner to have a new struct that stores both the `iconv_open`d converter and the bad byte handler so they can be passed around together. What do you think?

evanmiller:
I was thinking the extra handler would operate on the result of readstat_convert rather than inside it. In pseudo-code:

```
foreach (text) {
  if (good = readstat_convert(text)) {
   handle.value(good);
  } else {
   result = handle.bad_sequence(text);
   if (result == ABORT) {
     return
   } else {
     handle.value(text)
   }
  }
}
```

gorcha:
(Sorry for the delay getting back to this!)

Oh cool, that make sense. Thinking about it more we probably only need the bad byte handler for values as well, I haven't seen this come up with metadata at all.

I'll have another go and update the PR, thanks for the feedback!

skalteis:
Hi @evanmiller @gorcha,
are there any plans for getting this PR merged for the next release and subsequent inclusion in haven?
Sorry for nagging...  thank you for you work! ;-)

gorcha:
Hi @skalteis, thanks for the reminder! 🙂

I still need to redo this PR given Evan's feedback. It's a bit of work so haven't been able to find the time to sort it out just yet but hopefully will over the next couple of weeks.

PR #256, opened 2021-11-17
Fix unicode non-ASCII name check with unsigned char

pinotree:
In case 'char' is unsigned (which happens on some architectures), then
checks like `str[i] > 0` will always be true (for non-NULL characters,
of course). Since we are interested to check only for the values <= 127,
then change the check to cast the char to unsigned, so we can always
check that the value is lower than 0x80 (128).


evanmiller:
Can we simplify this logic with `(unsigned char)str[i] < 0x80`?

pinotree:
> Can we simplify this logic with `(unsigned char)str[i] < 0x80`?

Oh indeed; apparently I complicated this overlooking the easy solution; thanks!

Issue #257, opened 2021-11-19
SAS Transport writer format parsing bug


gorcha:
Hi!

Just noticed there's a bug in the format parser when writing SAS xport files.

`xport_write_variables` is using `sscanf` to pull the format name, width and decimals:
https://github.com/WizardMac/ReadStat/blob/f8731f89d581b9839cdbdd5dec256f78d3d788f1/src/sas/readstat_xport_write.c#L114

Because there are no spaces between the name and the other bits the whole string is being consumed as the format name by `%s`, and width and decimals are never set. In many cases this is still interpreted correctly from the name alone, but others have issues, for e.g.:
* COMMA10.3 - longer than 8 chars, so is cut off when storing the format name
* 10.3 - fails with an invalid name warning in SAS

It seems like a bit of a pain to parse - the format is `<name><width>.<decimal>`, but width and decimal and the period between them are all optional. Name follows the [SAS naming rules](http://documentation.sas.com/doc/en/lrcon/9.4/p18cdcs4v5wd2dn1q0x296d3qek6.htm), with the additional requirement that it can't end in a number.

evanmiller:
Thanks for the report. I think a format string like `"%[A-Z]%d.%d"` might do the trick, but this will need testing.

gorcha:
Yeah that was my initial thought as well, but unfortunately it's a bit more irritating - the name can contain numbers as well so it probably needs a simple FSM.

The regex for the format string would be something silly like `($?[A-Za-z_]([A-Za-z_0-9]*[A-Za-z_])?)([0-9]*\.?)?([0-9]*)?`

So numeric sequences need to take context into account. It seems easiest parsing from the end, so ignoring any validation of the format name itself something along these lines:
```c
int num_seq_end;
for (i = strlen(format) - 1; i > 0; i--) {
  switch(format[i]) {
    case '.':
      if (state == DECIMAL) { /* i + 1 to num_seq_end as decimal value */ }
      state = PERIOD;
      break;
    case '0'...:
      if (state == START) { state = DECIMAL; num_seq_end = i; }
      if (state == PERIOD) { state = WIDTH; num_seq_end = i; }
      break;
    default:
      if (state == DECIMAL) { /* i + 1 to num_seq_end as decimal value */ }
      if (state == WIDTH) { /* i + 1 to num_seq_end  as width value */ }
      state = NAME;
  }

  if (state == NAME) {
    /* start to i characters as format name */
    break;
  }
}
```

PR #258, opened 2021-11-21
Parse SAS format strings for XPT outputs (#257)

gorcha:
This PR adds a simple parser to extract format name, width and decimals from a SAS format string to be used when writing XPT files.

It parses the format string in reverse, assigns width and decimals based on the pattern of numbers and periods at the end, and uses the remainder of the string as the format name.

There are also a couple of tweaks to the section that writes the long format name for XPT v8 files:
* Now uses a simplified version of the parser to extract the format name
* Writes the entire format name to the long format record per the [spec](https://support.sas.com/content/dam/SAS/support/en/technical-papers/record-layout-of-a-sas-version-8-or-9-data-set-in-sas-transport-format.pdf)
* The long format record was incorrectly writing the label out after the format/informat values, these have been swapped.

### DIFF
--- a/src/sas/readstat_xport_write.c
+++ b/src/sas/readstat_xport_write.c
@@ -109,9 +109,60 @@ static readstat_error_t xport_write_variables(readstat_writer_t *writer) {
         if (variable->format[0]) {
             int decimals = 0;
             int width = 0;
-            char name[24];
 
-            sscanf(variable->format, "%s%d.%d", name, &width, &decimals);
+            char name[24];
+            name[0] = '\0';
+
+            char format_data[256];
+            memcpy(&format_data, variable->format, sizeof(variable->format));
+
+            /*
+             * STATES
+             * 0 - start
+             * 1 - first number (width or decimal depending on the presence of period)
+             * 2 - period
+             * 3 - second number (width)
+             * 4 - name
+             */
+            int state = 0;
+            int j;
+
+            for (j=strlen(format_data)-1; j>=0; j--) {
+                switch(format_data[j]) {
+                case '.':
+                    if (state == 1) {
+                        sscanf(&format_data[j + 1], "%d", &decimals);
+                    }
+                    state = 2;
+                    break;
+                case '0': case '1': case '2': case '3': case '4':
+                case '5': case '6': case '7': case '8': case '9':
+                    if (state == 0) {
+                        state = 1;
+                    } else if (state == 2) {
+                        state = 3;
+                        format_data[j + 1] = '\0';
+                    }
+                    break;
+                default:
+                    if (state == 1 || state == 3) {
+                        sscanf(&format_data[j + 1], "%d", &width);
+                        format_data[j + 1] = '\0';
+                    } else if (state == 2) {
+                        format_data[j + 1] = '\0';
+                    }
+                    state = 4;
+                }
+
+                if (state == 4) {
+                    sscanf(format_data, "%s", name);
+                    break;
+                }
+            }
+
+            if (state == 1 || state == 3) {
+                sscanf(format_data, "%d", &width);
+            }
 
             copypad(namestr.nform, sizeof(namestr.nform), name);
             namestr.nfl = width;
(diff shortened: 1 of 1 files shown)


gorcha:
I've made a few more updates after seeing the test failures. The XPT reader code has been updated to reflect the changes to the XPT writer.

I had to make a couple of changes to the tests for what looked like false positives, and I think they make sense:
* One of the SAS format roundtrip tests has been changed to use `$CHAR3` instead of `$CHAR3.`. The trailing period is dropped when roundtripping XPT files because of the way the format is reconstructed so `$CHAR3.` is read back in as `$CHAR3` and fails the test. It represents the same format so it made to sense to me to update the test.
* It now checks if label_set has been set before running the roundtrip test (as was already the case for format and display_width). It was throwing some strange errors in the display_width tests that looked like it was checking unallocated memory like:
```
Test "Display width" failed: Column label sets
 * Format: xpt5 (0x200000)
 * Expected: "¤~y²U" (length=6)
 * Received: "~y²U" (length=5)
 * Column: 1
```

src/sas/readstat_xport_write.c
```diff
@@ -173,31 +224,28 @@ static readstat_error_t xport_write_variables(readstat_writer_t *writer) {
...
-                int width = 8;
-
-                int matches = sscanf(variable->format, "%s%d.%d", format_name, &width, &decimals);
-                if (matches < 1) {
-                    retval = READSTAT_ERROR_BAD_FORMAT_STRING;
-                    goto cleanup;
+                int j;
+                for (j=format_len-1; j>=0; j--) {
+                  if (!((variable->format[j] >= 48 && variable->format[j] <= 57) ||
+                          variable->format[j] == 46)) {
```
evanmiller, line 237: Maybe reuse the parser from above? If not, please use the C literal syntax for characters, `'0'`, `'9'`, etc

evanmiller (changes requested):

src/sas/readstat_xport_write.c
```diff
@@ -109,9 +109,60 @@ static readstat_error_t xport_write_variables(readstat_writer_t *writer) {
...
+
+            /*
+             * STATES
+             * 0 - start
+             * 1 - first number (width or decimal depending on the presence of period)
+             * 2 - period
+             * 3 - second number (width)
+             * 4 - name
+             */
+            int state = 0;
```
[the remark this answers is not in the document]
gorcha: Shall do, thanks!

src/sas/readstat_xport_write.c
gorcha: Yep I think best to use the parser from above once it's set up. Thanks!

gorcha:
I've updated the format parsing code to use a Ragel parser and added an xport_format_t struct to store the parser output, and added a few more tests to check it's throwing errors correctly.

I've also had a go at the fuzzer setup but a bit less sure of what I'm doing there!

Makefile.am
```diff
@@ -95,6 +96,7 @@ noinst_HEADERS = \
        src/sas/readstat_sas.h \
        src/sas/readstat_sas_rle.h \
        src/sas/readstat_xport.h \
+	   src/sas/readstat_xport_parse_format.h \
```
evanmiller, line 99: Please fix the whitespace here

evanmiller (changes requested):
Looking good! Just found one whitespace issue, otherwise I'm happy with it.

Thanks for taking the time to implement the fuzzer – and of course the word dictionary is a nice touch.

evanmiller:
It looks like Appveyor is failing:

https://ci.appveyor.com/project/evanmiller/readstat/builds/41638815

You may need to modify the Visual Studio project file to ensure all the new files are being built. See

https://github.com/WizardMac/ReadStat/blob/master/VS17/ReadStat.vcxproj

Makefile.am
gorcha: Done!

evanmiller:
Thanks, much better now!

mstackhouse:
@evanmiller - any idea when these changes might be able to be merged into a new release of ReadStat? These changes would open up some [important doors](https://github.com/tidyverse/haven/issues/650) for haven, which will have a significant impact in some pharmaceutical industry projects.

PR #277, opened 2022-07-18
SAV reader: Ignore bad timestamps

gorcha:
This PR is a fix for tidyverse/haven#683 in line with the change to DTA timestamp reading that fixed tidyverse/haven#488.

If the timestamp string is invalid the parser will ignore it and move on instead of throwing an error.


gorcha:
Hi @evanmiller, just wondering if you've had a chance to look at this?

evanmiller:
Fixed in https://github.com/WizardMac/ReadStat/commit/428ba68c4d6c74d69a4a2811fda623487fa76c9e

gorcha:
Thanks!

buhrmann:
Hi, would you have a time frame for a new release containing this fix? I'm using ReadStat through pyreadstat, and I'm reluctant to ask them to update their package if there is no new official release of the upstream code.

PR #290, opened 2023-01-10
Fix SAS catalog parsing failure when formats have no labels

gorcha:
This PR fixes an "Unable to allocate memory" error when attempting to parse sas7bcat files containing a format with no value labels.

This fixes a couple of issues reported in haven:
tidyverse/haven#653
tidyverse/haven#705

Thanks!


evanmiller:
Thanks!

PR #291, opened 2023-01-11
Check for long names in u64 SAS catalog files

gorcha:
This PR adds support for long names in u64 SAS catalog files. It looks like the bit flag for u64 catalog files is different to others - I'm only working off a single example file so possibly a bit of a long shot but there's a single bit difference (B8 for long names vs 98 for short names in the example file) so it looks like 0x20 is the correct bit mask, and this loads the catalog file successfully.

From tidyverse/haven#680 - example catalog file at [sas_catalog_file.zip](https://github.com/tidyverse/haven/files/9062177/sas_catalog_file.zip)

Thanks!


evanmiller:
Seems plausible, let's try it out

PR #293, opened 2023-01-19
Fix bugs when reading big-endian u64 SAS catalog files

gorcha:
This PR fixes a few bugs when reading big-endian u64 SAS catalog files.

In some cases ReadStat is currently checking a single byte for a value which is actually stored as a 16 bit int, which works fine on little-endian files but returns 0 when a file is big-endian.

The other changes just change the integer size being read from the file - it was incorrect in a few spots but only causes an issue when it's a big-endian file.

From tidyverse/haven#696 - example catalog files at [haven_reprex_github.zip](https://github.com/tidyverse/haven/files/9585025/haven_reprex_github.zip). This zip contains the same format file produced by multiple versions of SAS, so I've been working off direct comparisons between the two files.

Thanks!


evanmiller:
Thanks!

PR #295, opened 2023-02-19
Add strict prototypes for functions with no arguments.

adriaandegroot:
Using compile flags `-Werror -Wstrict-prototypes` (with Clang at least), the "prototypes" of some functions are flagged as an error. This is because of the difference between:
```
	int no_args_please(void);
```
and
```
	int no_args_please();
```
Where the former is a strict prototype that says "no arguments are allowed". With the compile flags mentioned, ReadStat fails to compile with error messages like this:

```
In file included from src/readstat_bits.c:9:
src/readstat_bits.h:10:29: error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]
int machine_is_little_endian();
```

This PR adds the missing `void` in a handful of places where Clang complains about it with those strict compile flags.

### DIFF
--- a/src/readstat_bits.h
+++ b/src/readstat_bits.h
@@ -7,7 +7,7 @@
 #undef READSTAT_MACHINE_IS_TWOS_COMPLEMENT
 #define READSTAT_MACHINE_IS_TWOS_COMPLEMENT 0
 
-int machine_is_little_endian();
+int machine_is_little_endian(void);
 
 char ones_to_twos_complement1(char num);
 int16_t ones_to_twos_complement2(int16_t num);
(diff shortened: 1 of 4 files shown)


adriaandegroot:
As packager for ReadStat on FreeBSD, I have added these patches to the packaging-bits to make it compile on FreeBSD 14-CURRENT (which defaults to Clang + strict flags). I have not tested with other compilers, but the semantic difference between the two "prototypes" has been relevant since before C11 and apparently C11 makes the `no_args_please()` form obsolescent.

evanmiller:
Thanks!

PR #296, opened 2023-03-25
Fix some typos

goggle:


evanmiller:
Thanks for the submission.

Please note that `src/sas/readstat_xport_parse_format.c` is auto-generated from `src/sas/readstat_xport_parse_format.rl` so the source file will need to be updated.

goggle:
I'm not sure how this works...
But I could revert the changes to the file `src/sas/readstat_xport_parse_format.c` if that helps.

evanmiller:
Yes please revert those changes and I will take care of the rest.

PR #298, opened 2023-04-18
Fix use after free

gerlachs:
free(rs_ctx) is called before last use of rs_ctx.


pinotree:
@evanmiller would it be possible to take a look at this? It should fix an error when building with GCC 13, see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1037838. Thanks!

Issue #300, opened 2023-05-22
UCRT64 builds of readstat 1.1.9 fail


JorisGoosen:
Using RTools42 to compile on Windows gives the following error during compilation:
```
  CC       src/spss/libreadstat_la-readstat_zsav_read.lo
  CC       src/spss/libreadstat_la-readstat_zsav_write.lo
In file included from D:/rtools42/ucrt64/include/unistd.h:11,
                 from src/readstat_io_unistd.c:12:
D:/rtools42/ucrt64/include/process.h:196:28: error: conflicting types for 'spawnv'; have 'intptr_t(int,  const char *, char * const*)' {aka 'long long int(int,  const char *, char * const*)'}
  196 |   _CRTIMP intptr_t __cdecl spawnv(int,const char *_Filename,char *const _ArgList[]) __MINGW_ATTRIB_DEPRECATED_MSVC2005;
      |                            ^~~~~~
<command-line>: note: previous declaration of 'spawnv' with type 'intptr_t(int,  const char *, const char * const*)' {aka 'long long int(int,  const char *, const char * const*)'}
make: *** [Makefile:2096: src/libreadstat_la-readstat_io_unistd.lo] Error 1
make: *** Waiting for unfinished jobs....
In file included from D:/rtools42/ucrt64/include/unistd.h:11,
                 from D:/rtools42/ucrt64/include/zconf.h:488,
                 from D:/rtools42/ucrt64/include/zlib.h:34,
                 from src/spss/readstat_sav_write.c:25:
D:/rtools42/ucrt64/include/process.h:196:28: error: conflicting types for 'spawnv'; have 'intptr_t(int,  const char *, char * const*)' {aka 'long long int(int,  const char *, char * const*)'}
  196 |   _CRTIMP intptr_t __cdecl spawnv(int,const char *_Filename,char *const _ArgList[]) __MINGW_ATTRIB_DEPRECATED_MSVC2005;
      |                            ^~~~~~
<command-line>: note: previous declaration of 'spawnv' with type 'intptr_t(int,  const char *, const char * const*)' {aka 'long long int(int,  const char *, const char * const*)'}
make: *** [Makefile:2285: src/spss/libreadstat_la-readstat_sav_write.lo] Error 1
In file included from D:/rtools42/ucrt64/include/unistd.h:11,
                 from D:/rtools42/ucrt64/include/zconf.h:488,
                 from D:/rtools42/ucrt64/include/zlib.h:34,
[...]

PR #301, opened 2023-05-23
Add option to disable spawnv define

JorisGoosen:
Helps UCRT mingw users workaround the problem mentioned in https://github.com/WizardMac/ReadStat/issues/300

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -42,6 +42,17 @@ AC_SUBST([SANITIZERS])
 
 AM_ICONV
 
+AC_ARG_ENABLE(
+    [spawnvfix], 
+    [AS_HELP_STRING([--disable-spawnvfix], [Do not use _spawnv=spanv define on MINGW builds because it breaks UCRT builds.])]
+)
+
+use_spawnvfix=""
+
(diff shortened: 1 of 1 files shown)


JorisGoosen:
@evanmiller ?

evanmiller:
I would prefer to have the script detect the platform and do the right behavior, rather than add an option.

JorisGoosen:
> I would prefer to have the script detect the platform and do the right behavior, rather than add an option.

Fair enough, Ill have a look

JorisGoosen:
The new configure doesnt require a manual option anymore, I tested it on Mingw x64 and Mingw UCRT x64 and it now configures and compiles on both without a hitch

evanmiller:
There's probably a cleaner way to do this with autconf (see https://stackoverflow.com/questions/46704731/check-value-of-c-define-with-autoconf) but this seems good enough for now, thanks

JorisGoosen:
Ah yeah that SO answer looks cleaner for sure, but thanks for merging it anyway. It will make our build process much smoother.

Issue #285, opened 2022-11-22
Security: heap-buffer-overflow in readstat_convert


hedsnz:
Originally reported here: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=33991

Also listed as SONATYPE-2021-1261 and https://osv.dev/vulnerability/OSV-2021-732.

PR #303, opened 2023-07-26
Add a bound check for sas7bcat_parse_value_labels when finding the offset of labels.

### LINKED ISSUE #299: Security: Heap-buffer-overflow in sas_read2
Hi Evan!

Are you able to see this issue?
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=55308

Our internal fuzzer picked it up, but the external report might not be fully visible yet.

Let me know if there is more that I can do to help!

Thanks for everything!

kiwiwarmnfuzzy:
`lbp1` will try to chase and read as many value labels as `label_count_capacity`, which may result in accessing memory locations past the allocated region, especially for malformed data. This also results in Address sanitizer error on some inputs. Add a bounds check to avoid it.

Closes #299
Maybe related to #285

### DIFF
--- a/src/sas/readstat_sas7bcat_read.c
+++ b/src/sas/readstat_sas7bcat_read.c
@@ -42,7 +42,7 @@ static void sas7bcat_ctx_free(sas7bcat_ctx_t *ctx) {
     free(ctx);
 }
 
-static readstat_error_t sas7bcat_parse_value_labels(const char *value_start, size_t value_labels_len, 
+static readstat_error_t sas7bcat_parse_value_labels(const char *value_start, const char *value_end, size_t value_labels_len, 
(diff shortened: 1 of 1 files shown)


kiwiwarmnfuzzy:
@evanmiller Hi there! Please take a look when you have time - this addresses one of the vulnerabilities identified by our fuzzing tool; realistically, this can happen with incorrect or malformed data.

evanmiller:
Hi, I'm just confused because I thought line 63 was performing this check

```
        if (&lbp1[3] - value_start > value_labels_len || sas_read2(&lbp1[2], ctx->bswap) < 0) {
```

But maybe I'm missing something?

evanmiller:
Ah, looks like a regression introduced in #293. I will prepare a fix.

PR #309, opened 2024-01-11
handles metadata timezone differences in sas

SInginc:
Dear maintainers and reviewers,

- Previously, it is known that dataset creation/modification time is stored in the header of a SAS dataset. It is assumed that these 2 datetimes are saved as UTC timezone.
- After an investigation, it turns out that SAS will store local time instead of UTC time in the header under some circumstances. 

To address this:

1. Found the place where SAS store the timezone difference
2. Convert local time back to UTC time

This PR will:

1. Reads timezone difference (which is stored just after the creation/modification time place)
2. Convert local time to UTC time: 
    - Timezone difference is store as seconds between UTC and local time with sign. E.g., Eastern Standard time (UTC-5) is -18000.
    - So, local time minus timezone difference is UTC time (`local_time = UTC + diff`).

### DIFF
--- a/src/sas/readstat_sas.c
+++ b/src/sas/readstat_sas.c
@@ -124,7 +124,8 @@ static time_t sas_epoch(void) {
     return - 3653 * 86400; // seconds between 01-01-1960 and 01-01-1970
 }
 
-static time_t sas_convert_time(double time, time_t epoch) {
+static time_t sas_convert_time(double time, double time_diff, time_t epoch) {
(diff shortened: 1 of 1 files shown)


evanmiller:
If you remove the commented-out code, this should be good to merge. Thanks!

SInginc:
> If you remove the commented-out code, this should be good to merge. Thanks!

Hi @evanmiller , the commented-out code has been removed.

Thank you!

PR #311, opened 2024-02-09
Fix a buffer overflow

### LINKED ISSUE #285: Security: heap-buffer-overflow in readstat_convert

gaborcsardi:
It happens if raw_str_used underflows and ends up a very large number, which is then used as the size of a string.

Closes #285.

### DIFF
--- a/src/spss/readstat_sav_read.c
+++ b/src/spss/readstat_sav_read.c
@@ -717,7 +717,7 @@ static readstat_error_t sav_process_row(unsigned char *buffer, size_t buffer_len
             }
             if (++offset == col_info->width) {
                 if (++segment_offset < var_info->n_segments) {
-                    raw_str_used--;
+                    if (raw_str_used > 0) raw_str_used--;
                 }
                 offset = 0;
                 col++;


evanmiller:
I think this should return a parse error, assuming the test cases represent invalid input?

gaborcsardi:
Yes, you do get a parse error afaict: 
```
❯ readstat ~/Downloads/clusterfuzz-testcase-minimized-fuzz_format.sav out.csv
Failed to find VAR252
Failed to find VAR253
Failed to find VAZ255
Failed to find VAR256
Failed to find VAR252
Failed to find VAR253
Failed to find VAZ255
Failed to find VAR256
Converted 5 variables and 3 rows in 0.00 seconds
Error processing /Users/gaborcsardi/Downloads/clusterfuzz-testcase-minimized-fuzz_format.sav: Unable to convert string to the requested encoding (invalid byte sequence)
```

If this is what you meant.

evanmiller:
Cool thanks

PR #349, opened 2025-11-23
Fix XPT writing for LABELV8/9 headers

gorcha:
This PR updates writing for OBSV8 and LABELV8/9 headers in SAS XPORT files (see tidyverse/haven#784).

LABELV8/9 headers are only used when writing variable labels longer than 40 characters and includes the number of variables with long labels in the header, but this number is left aligned instead of right aligned.

The OBSV8 headers were working fine but I cleaned up the function a little while I was at it since the label and obs headers are now using their own functions for V8 files instead of sharing a header writing function due to the differing alignment.

Have tested with SAS and all looks to be working fine.

### DIFF
--- a/src/sas/readstat_xport_write.c
+++ b/src/sas/readstat_xport_write.c
@@ -49,12 +49,21 @@ cleanup:
     return retval;
 }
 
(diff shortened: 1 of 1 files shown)


evanmiller:
Thanks!

PR #351, opened 2025-12-23
Support non ASCII characters in MR Sets

gorcha:
From tidyverse/haven#788 - the new MR set changes only support ascii characters in the set name in the ragel parser, but SPSS uses the file code page/UTF-8 for these.

This PR updates the parser to allow for non-ASCII characters, and runs text from the MR set through `readstat_convert()` to make sure the character encoding comes in correctly.

### DIFF
--- a/src/spss/readstat_sav_parse_mr_name.rl
+++ b/src/spss/readstat_sav_parse_mr_name.rl
@@ -3,18 +3,25 @@
 #include <stdlib.h>
 #include "../readstat.h"
 #include "../readstat_malloc.h"
+#include "../readstat_iconv.h"
+#include "../readstat_convert.h"
+#include "readstat_sav.h"
 
 %%{
     machine mr_extractor;
 
     action extract_mr_name {
-        mr_name = (char *)readstat_malloc(p - start + 1);
+        size_t src_len = p - start;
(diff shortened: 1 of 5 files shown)


src/spss/readstat_sav_parse_mr_name.rl
```diff
@@ -92,7 +106,7 @@
         mr_subvariables[mr_subvar_count++] = subvar;
     }
 
-    nc = (alnum | '_' | '.' ); # name character (including dots)
+    nc = ([^ =]); # name character (all characters except space and equals)
```
evanmiller, line 109: This seems excessively lax?

gorcha: My thinking was that we're just using this for pattern matching to extract the components of the field rather than actually validating the contents so it doesn't matter too much if it lines up with actual allowed characters.

The attempt to peg it to actual allowed characters in the current version is what's causing the error in tidyverse/haven#788 (and it would be a massive pain to build a character class that matches all potential valid chars from UTF-8 and other encodings). Can tighten it up a bit if there are particular things you're worried about?

gorcha: Hey @evanmiller, just following up - keen to get this merged so we can pull the change into haven. Thanks!

gorcha:
Thanks @evanmiller!

PR #353, opened 2026-01-07
Use user width when writing long string value label records in SAV writer

### LINKED ISSUE #323: readstat strips value_labels from str columns
[string+value labels issue.zip](https://github.com/user-attachments/files/18538600/string%2Bvalue.labels.issue.zip)

I read the data for one column of str with value labels and write it back to sav. Reading the sav file seems to be correct, hoever
the output does not contain the value labels for this string column, as it was defined in the input.
so I assume that today readstat strips the value labels.

Can be recreated by reading the _in file and exporting the _out file.

(I was originally using [pyreadstat](https://github.com/Roche/pyreadstat) I got directed to this repo via [this issue](https://github.com/Roche/pyreadstat/issues/281#issuecomment-2598144336))

gorcha:
The SAV writer functions that emit value labels for long string variables assume that the variable storage width is wider than the value label keys, but this is not checked during file writing so ReadStat can produce invalid files that don't throw an error on write (see tidyverse/haven#537).

This PR initially added a check for this during writing, including a minor update to the test suite to support tests with a manually specified storage width. While I was testing I noticed that PSPP isn't happy with using the storage width if it is wider than the user width and you get an error like:
```
warning: `test-550.sav' near offset 0x18c: Ignoring long string value label record for variable a because the record's width (16) does not match the variable's width (10).
```

I've updated the long string value writer to use the user width instead of storage width, and confirmed files now load successfully in PSPP (and still roundtrip successfully in ReadStat).

I think this fixes #323, tidyverse/haven#550 and Roche/pyreadstat#264.

### DIFF
--- a/src/spss/readstat_sav_write.c
+++ b/src/spss/readstat_sav_write.c
@@ -913,6 +913,7 @@ static readstat_error_t sav_emit_long_string_value_labels_record(readstat_writer
         for (k=0; k<var_count; k++) {
             readstat_variable_t *r_variable = readstat_get_label_set_variable(r_label_set, k);
             int32_t name_len = strlen(r_variable->name);
+            int32_t user_width = r_variable->user_width;
             int32_t storage_width = readstat_variable_get_storage_width(r_variable);
             if (storage_width <= 8)
                 continue;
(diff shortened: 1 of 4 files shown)


evanmiller:
Thanks!

PR #354, opened 2026-01-07
Allow 0 width columns in sas7bdat reader

gorcha:
This PR fixes an error when reading SAS files with zero observations (see tidyverse/haven#627). When there are zero observations SAS appears to write string variables with a default width of 0, which was throwing an error in `sas7bdat_validate_column()`.

I've removed the check for width == 0 since this is allowed by SAS and the zero observation data file loads successfully so this check seems unnecessary.


evanmiller:
Thanks!

PR #360, opened 2026-03-17
Validate XPort v5 variable name length (max 8 chars)

andreasnoack:
The XPort writer was using sas_validate_variable which allows 32-char names, but v5 XPort only supports 8-char names (nname[8]). Names longer than 8 chars were silently truncated. Now v5 returns READSTAT_ERROR_NAME_IS_TOO_LONG instead.

(Commit created with Claude Code)


andreasnoack:
I believe the Windows failure is unrelated to the changes here

PR #361, opened 2026-03-26
Add support for w.d format

belegdol:
These appear to be stored without a specific name. Defined here: https://documentation.sas.com/doc/en/pgmsascdc/9.4_3.5/leforinforref/n1n7bmvs1brl23n1dj9cbzau39df.htm

### DIFF
--- a/src/sas/readstat_sas7bdat_read.c
+++ b/src/sas/readstat_sas7bdat_read.c
@@ -696,7 +696,7 @@ static readstat_variable_t *sas7bdat_init_variable(sas7bdat_ctx_t *ctx, int i,
         goto cleanup;
     }
     size_t len = strlen(variable->format);
-    if (len && ctx->col_info[i].format_width) {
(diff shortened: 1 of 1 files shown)


src/sas/readstat_sas7bdat_read.c
```diff
@@ -696,7 +696,7 @@ static readstat_variable_t *sas7bdat_init_variable(sas7bdat_ctx_t *ctx, int i,
         goto cleanup;
     }
     size_t len = strlen(variable->format);
-    if (len && ctx->col_info[i].format_width) {
+    if ((len && ctx->col_info[i].format_width) || (len == 0 && ctx->col_info[i].format_width)) {
```
hpoettker:
```suggestion
    if (ctx->col_info[i].format_width) {
```
The condition on `len` seems redundant when `len` may either be non-zero or zero,

belegdol: Good point, thanks for the review!

Issue #363, opened 2026-04-01
format lengths and digits are sometimes not read correctly


belegdol:
Hello,

while working on https://github.com/WizardMac/ReadStat/pull/361 I have noticed that format lengths and labels are sometimes not read at all - even for the formats not needing the PR. First investigation seems to indicate that it might have something to do with data representation: `SOLARIS_X86_64, LINUX_X86_64, ALPHA_TRU64, LINUX_IA64` works fine but `WINDOWS_64` does not.

belegdol:
This happens because `format_width` and `format_digits` are only read for files for which `ctx->u64` is `true`, but for `WINDOWS_64` files it is `0`.

belegdol:
Fixed in 5c446781f5ac7d716fae0ad0ef4a3f3920eb2cd6.

PR #364, opened 2026-04-01
Read format_width and format_digits for 32-bit files

belegdol:
Tested using a file created on Windows (WINDOWS_64).
This appears to fix issue #363.


evanmiller:
Thanks!

PR #350, opened 2025-11-24
test: fix build test workflow on macOS

shun2wang:
also Linux build dependencies were cleaned up.

Don't know why Windows build still failed.

PR #342, opened 2025-07-16
Drop gettext as no longer needed

sarahec:
> error: possibly undefined macro: AM_ICONV
>	    If this token and others are legitimate, please use m4_pattern_allow.
>	    See the Autoconf documentation.

This happens with gettext 0.25. 

`gettext` is there to supply `iconv` in older scripts. `iconv()` has been obsolete since 2001.

Fixes #341

PR #367, opened 2026-04-19
Fix Windows pipeline

hpoettker:
This PR fixes the Windows build pipeline, which is currently failing.

### The make step

The make step is currently failing on Windows as the linker complains about unresolved references to `libiconv` functions.

The proposed fix is to replace the variable `LTLIBICONV` in `configure.ac` by hard-coded values:

- `-liconv` for the Windows build
- `''` for the other builds

The variable `LTLIBICONV` should actually be created by the macro `AM_ICONV` but this was removed in #342. And the dependency `gettext` that is needed for `AM_ICONV` was removed in #350.

This is probably not the cleanest fix but it keeps changes restricted to the Windows build.

### The make check step

When the `make` step is fixed in the Windows pipline, the `make check` currently fails as the next step.

The problem is that the compiler is missing the flag `-D_spawnv=spawnv`.

This is caused by the change from #301.

The proposed change works as expected in the pipeline. And it should also work for UCRT.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -54,20 +54,15 @@ AS_CASE([$host],
 AC_SUBST([EXTRA_WARNINGS])
 
 AS_CASE([$host],
-	[*mingw*|*cygwin*], [EXTRA_LDFLAGS="$LTLIBICONV -no-undefined"],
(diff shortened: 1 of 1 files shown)


evanmiller:
Thanks!

PR #394, opened 2026-09-12
Fix build

hpoettker:
Addresses the warning that "output may be truncated before the last format character."


evanmiller:
Thanks!

PR #396, opened 2026-09-20
Fix build in Windows C Runtime

hpoettker:
In the Windows C Runtime, the POSIX function `localtime_r`, which is a thread-safe alternative to `localtime`, is not available.

It offers the function `localtime_s` instead, which differs from the C11 function of that name in both parameter order and return value.


hpoettker:
Sorry for the multiple pushes.

The AppVeyor check is now successful as intended. In the commit amendments I tried to fix the artifact in the logs of the check that highlights an intended write to `stderr` of `ReadStat_Test_sav_date`. But that went nowhere and all changes to `appveyor.yml` are now reverted. The current commit contains the same changes as the first one.

